PR #3, opened 2025-03-07
Some feature updates

thomaskf:
1. Update the workflow so that the compiled IQ-TREE file under MacOS will not depend on the system local openmp library
2. Update the codes to remove the warnings during the compilation
3. Add the option --fundi-epsilon to set the epsilon value for funDi model
4. mAIC is disabled for -S option


bqminh (approved):
thanks look good

PR #5, opened 2025-03-12
Fixed the wall-clock time and allow predefined frequency for +F

thomaskf:
1. Fixed the wall-clock time issue
2. Allow predefined frequency for +F


bqminh (approved):
all good now thanks

PR #10, opened 2025-04-04
Naser-Khdour et al. (2024) -> Naser-Khdour et al. (2019)

HS6986:
Correct the publication year of [“The Prevalence and Impact of Model Violations in Phylogenetic Analysis”](https://doi.org/10.1093/gbe/evz193) to 2019, which was incorrectly stated as 2024


bqminh (approved):
thanks for the correction

PR #23, opened 2025-05-02
Updated ci.yaml to fix the issue of the Linux Intel binary file compiled on Github.

thomaskf:


bqminh (approved):
interesting to see that gcc-9 generates faster binary than gcc-12.

PR #24, opened 2025-05-02
Fix issue for MCMCTree dating when there is a leaf at first child of the super tree and Formatted code for Hessain calculation.

piyumalanthony:
Fix issue for MCMCTree dating when there is a leaf at first child of the super tree.
Formatted code for Hessain calculation for MCMCTree.


bqminh (approved):
a large block of new code in phyloanalysis.cpp at line 2943, which I couldn't review, but hope that's OK.

PR #29, opened 2025-05-10
Support Exporting Model and Partition Selections to MrBayes

IntegerLimit:
## Introduction

This PR ports over https://github.com/iqtree/iqtree2/pull/267 into IQTree 3, as well as implementing all review requests previously on that PR.

Similarly, https://github.com/iqtree/iqtree2/issues/195 has been implemented for all MrBayes supported sequence types. (DNA/RNA, Protein, Binary, Morphological & Codon) This PR is complete, excluding any changes required further from suggestions and bug fixes after testing has been conducted.

## General Implementation Details
- MrBayes files will be outputted when the user provides the `-mset mrbayes`, or adds the `-mrbayes` flag
- One MrBayes file will be exported: `.mr_bayes.nex`)
- MrBayes files will be outputted on partitioned and non-partitioned runs
- For non-supported sequence types and non-supported models (Mixture and Non-Reversible Models), a default of `GTR+G+I` (DNA) will be used, with warnings printed to log and file.
- `+R` has been mapped to `+G+I` (with warnings printed to log + file)

## DNA Fallbacks
For DNA, MrBayes supports three models: JR/F81 (`nst=1`), HKY (`nst=2`) and GTR (`nst=6`) (excluding their fixed frequency counterparts)

Therefore, when a model is used that is not supported in MrBayes, it will default to GTR, due to the lower impact of increased parameters when using Bayesian Inference.

## Protein Fallbacks
For Protein, when a model is used that is not supported by MrBayes, a default of `GTR` will be used. Then there will be a rate and state frequency matrix of the model included. The rate matrix will be set to `fixed`, unless the model used by IQTree was `GTR20`, in which case `dirichlet` will be used. This appears to be a mandatory parameter for MrBayes GTR models.

## Binary, Morphological and Codon Data Exclusions
- MrBayes Binary Data does not support +I, that has been ignored, with warning printed to log and file
- Morphological Data:
  - MrBayes does not support states {A-Z}, warning has been printed to log and file
  - MrBayes does not support +I, that has been ignored, with warning printed to log and 
- MrBayes Codon Data does not support any Heterogeneity Modifier (+G, +I or +R). They have been excluded, and a warning is printed to the log and file when a model is used with any modifiers.

### Codon Models in MrBayes: Introduction
The basic structure for codon models in MrBayes is quite similar to *mechanistic* codon models in IQTree, following the same formulation of the model by [Goldman & Yang 1994](http://mbe.oxfordjournals.org/content/11/5/725.abstract) and [Muse & Gaut 1994](http://mbe.oxfordjournals.org/content/11/5/715.abstract). However, the settings for MrBayes are under different names, and most inputs cannot be ported directly, making it the most difficult model to port to MrBayes format.

Instead of using named models, such as `MG` or `GY`, MrBayes uses five main parameters:
- Nucleotide Substitution Model (From `JC`, `HKY` and `GTR`), set through `lset nst` (`nst = 1` for `JC`, `nst = 2` for `HKY`, `nst = 6` for `GTR`)
- The Transition/Transversion (ts/tv) Rate Ratio (set through `prset tratiopr`)
- The Non-Synonymous/Synonymous (dn/ds) Rate Ratio (set through `prset omegapr`)
- The State Frequencies of each *non-stop* codon (set through `prset statefreqpr`)
- The Omega Variation Type and Corresponding Parameters (set through `lset omegavar`)

(Source: [MrBayes Manual (Chapter 6.1.3 & Appendix A)](https://github.com/NBISweden/MrBayes/blob/develop/doc/manual/Manual_MrBayes_v3.2.pdf), MrBayes `help lset` and `help prset` commands, [IQTree Documentation on Substitution Models (Section on Codon Models)](http://www.iqtree.org/doc/Substitution-Models#codon-models))

#### Nucleotide Substitution Model
For retrieving the nucleotide substitution model that should be used as input into MrBayes, the implemented code does the following:
- If `fix_kappa` is true, then the model will be set to `nst = 1` (JC)
- If `fix_kappa` is false, then the model will be set to `nst = 2` (HKY)

This implementation means that GTR is not used, appropriate considering the inputs for Mechanistic Codon Models in IQTree (ds/dt ratio + ts/tv ratio).

Note that `fix_kappa` is only set to true under the Codon Models `MGK` and `GY0K` (which are the only models without a `ts/tv` input ratio). This can be shown through the [`initCodon` function](https://github.com/iqtree/iqtree2/blob/master/model/modelcodon.cpp#L323-L366), which only calls `initMG94` or `initGY94` with `fix_kappa` as true for those two models. That input is then read into the `fix_kappa` field [here for MG Models](https://github.com/iqtree/iqtree2/blob/master/model/modelcodon.cpp#L515) and [here for GY Models](https://github.com/iqtree/iqtree2/blob/master/model/modelcodon.cpp#L551).

### Empirical Model Output
MrBayes does not support Empirical Codon Models, so when such a model is being used (or a mixture of Empirical + Mechanistic), a warning is printed to the log and file. However, a model is still outputted, with `nst = 6` (GTR-like model).

### Codon Codes
Whilst IQTree uses Number IDs for its Codon Codes (CODON1, CODON2, etc.), MrBayes uses Text IDs. (`vertmt`, `invermt`, etc.) There is no clear documentation or description for most of the models, but below shows the final table to transfer from IQTree Codon Codes to MrBayes Codon Codes.

An `XXX` in the MrBayes column represents a code that MrBayes does not support.

| IQTree  | MrBayes    |
| ------- | ---------- |
| CODON1  | universal  |
| CODON2  | vertmt     |
| CODON3  | yeast      |
| CODON4  | mycoplasma |
| CODON5  | invermt    |
| CODON6  | ciliate    |
| CODON9  | echinoderm |
| CODON10 | euplotid   |
| CODON11 | universal  |
| CODON12 | XXX        |
| CODON13 | XXX        |
| CODON14 | XXX        |
| CODON16 | XXX        |
| CODON21 | XXX        |
| CODON22 | XXX        |
| CODON23 | XXX        |
| CODON24 | XXX        |
| CODON25 | XXX        |

If a code is used that MrBayes does not support, it defaults to the universal code, and prints a warning to the log and file.


bqminh (approved):
all good after testings, thanks Thomas

PR #35, opened 2025-05-13
[WIP: Implementation of a New Algorithm] The extension of MixtureFinder to morphological data

HS6986:
Dear All,

**This is a draft pull request that extends a pull request in progress (https://github.com/iqtree/iqtree3/pull/11) and implements a new MixtureFinder-like algorithm for selecting the best-fitting mixture models for morphological data, which has not yet been validated in publications.** See below for details.

We're moving forward with the plan to extend MixtureFinder ([Ren et al., 2025](https://doi.org/10.1093/molbev/msae264)), which currently only works on DNA data, to codon, binary, and non-morphological multistate data in https://github.com/iqtree/iqtree3/pull/11. Morphological data are not considered in this PR because they are fundamentally incompatible with the existing MixtureFinder framework as they have a number of properties that differentiate it from other major data types:

- the number of states is different depending on the character
- the state labels are arbitrarily and different depending on the character
- they are artificially sampled so that they will completely or almost lack invariant (or sometimes parsimony uninformative) characters

These properties have led empiricists to use distinctive analytical conditions for morphological data in probabilistic phylogenetic methods. Generally:

- they partition data by the number of states (see [Černý & Simanoff, 2023](https://doi.org/10.1038/s41598-023-35784-3); https://davidcerny.github.io/post/teaching_revbayes/; https://github.com/davidcerny/GEOS26100-Fall2022; [Huang, 2025 preprint](https://doi.org/10.1101/2025.04.22.650124); https://github.com/ej91016/MorphoParse)
- they apply the MK models ([Lewis, 2001](https://doi.org/10.1080/106351501753462876)), which are models with equal state frequencies and replacement rates, thus avoiding making *a priori* assumptions of them
- they apply ascertainment bias corrections (e.g., [Lewis, 2001](https://doi.org/10.1080/106351501753462876)) to account for the lack or paucity of invariant (or parsimony uninformative) characters, after removing invariant (or parsimony uninformative) characters if they are present

However, I've recently learned that some Bayesian phylogenetic software (MrBayes and RevBayes) implement methods that model the heterogeneity of state frequencies among characters in morphological data using mixture models ([Wright et al., 2016](https://doi.org/10.1093/sysbio/syv122); https://revbayes.github.io/tutorials/morph_tree/). These methods do not seem to be widely used as far as I know (probably), but it may open up new avenues in morphological phylogenetics.

What I'm thinking is that the idea of modeling the heterogeneity of state frequencies (and perhaps also replacement rates) in morphological data using mixture models could be extended to maximum likelihood frameworks. In addition, a feature for automatic model selection in IQ-TREE similar to MixtureFinder could improve model fit for morphological data. I think it would be valuable, given that the aforementioned software do not implement such a feature.

I've implemented my devised MixtureFinder-like algorithm for selecting the best-fitting models for morphological data in this PR, although they have several limitations that promote further development: they currently cannot explicitly consider the heterogeneity of the numbers of states among characters (users probably need to test models per partition) and ascertainment bias corrections (`+ASC` in IQ-TREE) cannot be applied, as `+ASC` in mixture models is currently not implemented (https://github.com/iqtree/iqtree3/issues/12).

Of course, this new algorithm should be theoretically explained well and empirically validated in a peer-reviewed paper (or at least in a preprint) in the future before it is possibly merged into the master branch and explained in the documentation, I create this PR for now to potentially get some feedback.

I apologize for the current dirtiness of the code.

I'll post details of the algorithm, the usage details, and some test runs with empirical datasets later. I'm sorry, but it might take a few days or more.

If I have misunderstood something, or if this algorithm is fundamentally not justified in the first place, I apologize.


HS6986:
I'm sorry, but I realized that this algorithm doesn't seem to be justified at least in the current implementation. I apologize the confusion I may have caused. I am closing this PR for now.

PR #41, opened 2025-05-20
Urgent fix for MCMCTree control file generation and Hessian calculation with ASC flag.

piyumalanthony:
Urgent fix for MCMCTree control file generation and Hessian calculation with ASC flag


bqminh (approved):
looks good

PR #71, opened 2025-06-27
Add testing to workflow

thomaskf:
This testing procedure will verify the codes on 15 DNA datasets and 14 Amino Acid datasets under Linux Intel, Linux ARM, Mac Intel, Mac ARM and Windows Intel platforms.
This procedure also examines the corresponding running times.


bqminh (approved):
great, thanks!

PR #75, opened 2025-07-03
Add memory usage check to GitHub Actions workflow

Hashara:


bqminh (commented):
change memory-arg and runtime-arg to column name instead of column ID, easier to read and more general.
Use 'real' time to measure wall-clock time, not user time or log file of iqtree.
Document verify_runtime.sh and verify_memory.sh, what arguments do they take as input.

bqminh (approved):
thanks!

PR #76, opened 2025-07-07
fix: handle taxa count divisible by 32

magnified103:
The following snippet crashes under `-fsanitize=undefined` when the number of taxa is a multiple of 32.

https://github.com/iqtree/iqtree3/blob/7cded0cbc1767dab82bed39720a809f9359d7227/pda/split.cpp#L179-L185

The fix simply sets `num_bits` to 32 in that case. It's similar to the fix in https://github.com/iqtree/iqtree2/pull/412. I also manually checked all relevant parts in this file using the same logic.

I think we also need to backport this fix to IQ-TREE 2.


bqminh (approved):
thanks for discovering this!

PR #88, opened 2025-09-09
Bug fixed: delay deleting the "first_insertion" pointer to avoid crashing in partition simulation

trongnhanuit:
Bug fixed: delay deleting the "first_insertion" pointer to avoid crashing in partition simulation.


bqminh (approved):
all good

PR #93, opened 2025-09-23
Add option to fix branch lengths to those in the input tree when using fit_tree in libiqtree

thomaskf:
This pull request addresses issue: https://github.com/iqtree/piqtree/issues/311

The `fit_tree` function interface in libiqtree has been updated to include a new boolean parameter. This parameter allows users to specify whether the branch lengths should remain fixed to the values provided in the input tree, when a fixed topology is used.

This change allows users more control over tree fitting, ensuring that branch lengths can either be preserved as given or re-estimated as needed.


bqminh (approved):
all good thanks

PR #104, opened 2025-10-23
Update AliSim API in piqtree

trongnhanuit:
This pull request includes:
- Automatically convert a negative random seed into a positive one to ensure deterministic output.
- Consider more model names/aliases (provided by Robert McArthur) for detecting sequence type from the model name. This update is only applied when running AliSim, to ensure it won't affect other features in IQ-TREE.


bqminh (approved):
all good

PR #147, opened 2026-04-21
Update CI to compare the new IQ-TREE 3 vs IQ-TREE 2 (rather than pre-defined fixed values)

trongnhanuit:
This pull request updates the CI workflow to compare runtime and memory usage between IQ-TREE 3 and IQ-TREE 2, replacing the previous approach based on fixed pre-defined values.

Notes: Some tolerance thresholds have been adjusted to account for variability in runtime and memory usage across different runs.


bqminh (approved):
good now, thanks

PR #149, opened 2026-04-24
fix a bug: when +ASC model is applied

HuaiyanRen:
when +ASC model is applied, DNA model selection nested network crash.

issue #143


bqminh (approved):
looks good

bqminh:
Can you check why this PR failed the memory test: https://github.com/iqtree/iqtree3/actions/runs/24956817458/job/73076437679

PR #155, opened 2026-05-05
clamped_uint.cpp: Fix duplicate instantiation errors with Clang 21

Jehops:
The build was failing on FreeBSD/main after the import of clang 21.

Remove redundant explicit instantiations of template function in terraphast/lib/clamped_uint.cpp.  The preceding `template class` lines already instantiate the member functions, and the free function templates (operator== and operator!=) are defined in the same translation unit and will be instantiated on demand.

Clang 21 now rejects these as duplicate definitions.


bqminh (approved):
thanks for the fix

PR #158, opened 2026-05-06
Codon Newton fix and switch to Clang and embedded OpenMP for Linux

thomaskf:
This PR makes the following changes:

- Resolved an issue in the Newton optimization step for the codon model.
- Linux builds now use the embedded OpenMP library (LLVM libomp.a bundled in the source tree) instead of the system runtime.
- The GitHub Actions Linux workflow has been switched from GCC to Clang.
- Clang's link-time optimization (LTO) produces a special object-file format that the default Linux linker (ld) does not understand. lld (LLVM's linker) is used for the link step.
- Increased the memory-usage thresholds for some benchmark test cases to accommodate the additional usage introduced by the new mAIC calculation.

**Performance**

Switching to Clang together with the embedded OpenMP runtime produces a substantial speedup on Intel Linux. Benchmark wall-clock times measured in the GitHub Actions workflow:

| Compiler | IQTREE2 | IQTREE3 (before) | IQTREE3 (after) |
| --- | --- | --- | --- |
| GCC-9 | 6m57s | 6m56s | -- |
| GCC-12 | 6m08s | 6m16s | -- |
| Clang-13 | 7m30s | -- | 4m18s |
| Clang-15 | 6m55s | -- | 4m10s |

Note: The variation across rows reflects runner noise rather than a real change.


bqminh (approved):
very good, thanks

PR #168, opened 2026-05-21
Bug fix for PMSF+I combinations

### LINKED ISSUE #58: PMSF Step 1: '+I' Proportion of Invariable Sites Not Correctly Estimated in IQ-TREE v3.0.1
Dear IQ-TREE Developers,

I am conducting phylogenetic analysis using the PMSF model in IQ-TREE v3.0.1, with the model specified as LG+C10+I+G4. However, during the first step of the PMSF procedure, it seems that the proportion of invariable sites for the `+I` model was not correctly estimated.

`Estimate model parameters (epsilon = 0.010)
Thoroughly optimizing +I+G parameters from 10 start values...
Init pinv, alpha: 0.000, 1.000 / Estimate: 0.000, 0.374 / LogL: -1941771.816
Init pinv, alpha: 0.042, 1.000 / Estimate: 0.000, 0.374 / LogL: -1941771.848
Init pinv, alpha: 0.084, 1.000 / Estimate: 0.000, 0.374 / LogL: -1941771.842
Init pinv, alpha: 0.126, 1.000 / Estimate: 0.000, 0.374 / LogL: -1941771.842
Init pinv, alpha: 0.169, 1.000 / Estimate: 0.000, 0.374 / LogL: -1941771.842
Init pinv, alpha: 0.211, 1.000 / Estimate: 0.000, 0.374 / LogL: -1941771.845
Init pinv, alpha: 0.253, 1.000 / Estimate: 0.000, 0.374 / LogL: -1941771.849
Init pinv, alpha: 0.295, 1.000 / Estimate: 0.000, 0.374 / LogL: -1941771.846
Init pinv, alpha: 0.337, 1.000 / Estimate: 0.000, 0.374 / LogL: -1941771.843
Init pinv, alpha: 0.379, 1.000 / Estimate: 0.000, 0.374 / LogL: -1941771.845
Optimal pinv,alpha: 0.000, 0.374 / LogL: -1941771.816`

The dataset I used contains 86,625 sites, including 32,846 constant sites. The command I used is as follows:
`./iqtree3 -s cyanobacteria/alignment.nex -ft cyanobacteria/MF.treefile -m LG+C10+I+G4 -safe -pre cyanobacteria/infer/LG-C10fix-PMSF -n 0 -T 40
`

However, the `+G4` model seems to have been correctly estimated, while the results for the `+I` model appear to be unusual.

Thank you very much for your time and support.

StefanFlaumberg:
This pull request fixes #58 by using site-specific state frequencies to compute `ptn_invar` for invariant patterns in the `PhyloTree::computePtnInvar()` function under the PMSF model.

### DIFF
--- a/model/modelset.cpp
+++ b/model/modelset.cpp
@@ -125,6 +125,16 @@ void ModelSet::writeInfo(ostream& out)
 	}
 }
 
+void ModelSet::getStateFrequency(double *state_freq, int mixture) {
+    ASSERT(mixture >= -1);
+    if (mixture >= 0) {
+        at(mixture)->getStateFrequency(state_freq);
+        return;
+    }
+    // default: return the +F freqs across all patterns
+    ModelMarkov::getStateFrequency(state_freq);
+}
+
 void ModelSet::decomposeRateMatrix()
 {
     if (empty()) {
--- a/model/modelset.h
+++ b/model/modelset.h
@@ -130,6 +130,13 @@ public:
 	*/
 	virtual void writeInfo(ostream &out);
 
+    /**
+     *  Compute the state frequency vector
+     *  @param mixture Submodel id of the state frequency vector, 0 for default frequencies
+     *  @param[out] state_freq State frequency vector, must have the size of num_states
+     */
+    virtual void getStateFrequency(double *state_freq, int mixture = 0);
+
 	/**
 		decompose the rate matrix into eigenvalues and eigenvectors
 	*/
--- a/model/modelset.h
+++ b/model/modelset.h
@@ -135,7 +135,7 @@ public:
      *  @param mixture Submodel id of the state frequency vector, 0 for default frequencies
      *  @param[out] state_freq State frequency vector, must have the size of num_states
      */
-    virtual void getStateFrequency(double *state_freq, int mixture = 0);
+    virtual void getStateFrequency(double *state_freq, int mixture = -1);
 
 	/**
 		decompose the rate matrix into eigenvalues and eigenvectors
--- a/tree/phylotreesse.cpp
+++ b/tree/phylotreesse.cpp
@@ -547,68 +547,71 @@ void PhyloTree::computePtnFreq() {
 }
 
 void PhyloTree::computePtnInvar() {
-	size_t nptn = aln->getNPattern(), ptn;
-	size_t maxptn = get_safe_upper_limit(nptn)+get_safe_upper_limit(model_factory->unobserved_ptns.size());
-  // For PoMo, only consider monomorphic states and set nstates to the number of
-  // states of the underlying mutation model.
-	int nstates = model->getMutationModel()->num_states;
-    int x;
-    // ambiguous characters
+    size_t nptn = aln->getNPattern();
+    size_t maxptn = get_safe_upper_limit(nptn)+get_safe_upper_limit(model_factory->unobserved_ptns.size());
+    // set nstates:
+    // - for PoMo: to the number of monomorphic (boundary) states
+    // - for other models: to the number of states of the model
+    int nstates = model->getMutationModel()->num_states;
+    // protein ambiguous characters
     int ambi_aa[] = {
-        4+8, // B = N or D
-        32+64, // Z = Q or E
-        512+1024 // U = I or L
+        4+8,     // B = N or D
+        32+64,   // Z = Q or E
+        512+1024 // J = I or L
     };
-
+    // set state_freq:
+    // - for PoMo: to the stationary freqs of monomorphic (boundary) states
+    // - for SSF: to the site-specific stationary freqs, for each pattern below
+    // - for other models: to the stationary freqs of the model
     double state_freq[nstates];
-
-    // -1 for mixture model
-
-    // Again for PoMo, the stationary frequencies are set to the stationary
-    // frequencies of the boundary states.
     model->getMutationModel()->getStateFrequency(state_freq, -1);
-
-	memset(ptn_invar, 0, maxptn*sizeof(double));
-	double p_invar = site_rate->getPInvar();
-	if (p_invar != 0.0) {
-		for (ptn = 0; ptn < nptn; ptn++) {
-            if ((*aln)[ptn].const_char > aln->STATE_UNKNOWN)
+    memset(ptn_invar, 0, maxptn*sizeof(double));
+    double p_invar = site_rate->getPInvar();
+    if (p_invar) {
+        for (size_t ptn = 0; ptn < nptn; ++ptn) {
+            if (model->isSiteSpecificModel()) {
+                model->getStateFrequency(state_freq, ptn);
+            }
+            int cstate = aln->at(ptn).const_char;
+            if (cstate > aln->STATE_UNKNOWN) {
+                // var pattern or PoMo polymorphic const pattern
                 continue;
-
-			if ((*aln)[ptn].const_char == aln->STATE_UNKNOWN) {
-				ptn_invar[ptn] = p_invar;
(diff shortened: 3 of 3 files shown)


StefanFlaumberg:
**Two important notes:**
1. The `PhyloTree::computePtnInvar()` function still fails for PoMo polymorphic constant patterns (such as `0,0,1,1   0,0,1,1   0,0,1,1   0,0,1,1` for 4 species). Despite being constant, these patterns should be treated as variant patterns, since being polymorphic implies non-zero substitution rate. This is what both the old and the new comments say in the function. However, the `Alignment::computeConst()` function treats such patterns as constant, and hence they end up in the `else { ASSERT(0); }` case of `PhyloTree::computePtnInvar()`.
I haven't touched this issue here, because I'd rather address it in a comprehensive refactoring of how the `Alignment` class treats states (WIP for now).
2. The `PhyloTree::computePtnInvar()` function treats the unobserved patterns (`+ASC`) as gap-only/maximum uncertainty patterns, assigning them lh = 1 under the invariant category (they are among dummy patterns getting `ptn_invar[ptn] = 1 * p_invar`). **Is this the correct behaviour?**
Anyway, this is also a more complex issue, since `Alignment::computeConst()` is never called for the unobserved patterns and the behaviour of `PhyloTree::computePtnInvar()` should be consistent with that of `phylokernelnew.h`.

bqminh:
Claude AI review:

# Code Review ‚Äî PR #168 "Bug fix for PMSF+I combinations"

**Author:** @StefanFlaumberg ¬∑ **1 commit, +71/-51, 3 files ¬∑ mergeable: True / unstable ¬∑ 5/7 CI green, 2 platforms failing**

## Overview
Fixes issue #58: under PMSF (or any site-specific-frequency model) with `+I`, the proportion of invariable sites was estimated as 0. Root cause: `PhyloTree::computePtnInvar()` (`tree/phylotreesse.cpp:548`) computed each invariant pattern's contribution using a *single global* state-frequency vector instead of the per-pattern site-specific frequencies. The fix queries the per-pattern frequencies inside the loop when `model->isSiteSpecificModel()` is true, plus adds a missing `ModelSet::getStateFrequency` override so `ModelSet` (the SSF model) can hand out submodel/pattern-specific freqs.

## Correctness analysis

### The fix itself ‚Äî correct

- `tree/phylotreesse.cpp:573` adds `if (model->isSiteSpecificModel()) model->getStateFrequency(state_freq, ptn);` inside the per-pattern loop. This brings PMSF's per-pattern frequencies into the `ptn_invar` calculation. Without it, `ptn_invar[ptn] = p_invar * state_freq[const_char]` was using the +F averaged frequencies for every pattern ‚Üí pinv likelihood derivative was effectively flat and the optimiser parked it at zero. The math now matches what's expected for site-heterogeneous frequency models.
- `model/modelset.cpp:128` (`ModelSet::getStateFrequency(buf, mixture)`):
  - `mixture >= 0` ‚Üí returns `at(mixture)`'s freq (per-pattern submodel)
  - `mixture == -1` ‚Üí falls through to `ModelMarkov::getStateFrequency` (the +F average)
  This mirrors `ModelMixture::getStateFrequency` (`model/modelmixture.cpp:3806`) and is internally consistent.
- The outer `model->getMutationModel()->getStateFrequency(state_freq, -1)` call at `phylotreesse.cpp:572` is preserved, so for non-SSF models nothing changes ‚Äî the per-pattern call inside the loop is gated on `isSiteSpecificModel()`. ‚úì
- Comment cleanup, variable renaming (`cstate` ‚Üí `astate` for the ambiguity-decoded state), and tighter loop scoping are pure readability ‚Äî no behavioural change.

### One latent behavioural change worth flagging

`ModelSet::getStateFrequency` is **new**, with default `mixture = 0`. Before this PR, calling `modelset_ptr->getStateFrequency(buf)` (no mixture arg) fell through to `ModelMarkov::getStateFrequency`, which ignores `mixture` and returns the averaged `state_freq` member. **After this PR**, the same call dispatches to the new override and returns submodel **0**'s frequency, i.e. the *first pattern's* per-site frequencies ‚Äî not the average.

That matches the existing `ModelMixture` convention (same default-arg footgun there), so the PR isn't introducing a new inconsistency, but it *is* a silent behaviour change for any in-tree caller that does `m->getStateFrequency(buf)` on a ModelSet object without specifying `mixture`. I found ~20 such call sites:

- `main/phyloanalysis.cpp:656, 698, 4413, 4441`
- `model/partitionmodel.cpp:433`
- `model/modelfactory.cpp:653`
- `simulator/alisimulator.cpp:660, 691`
- `tree/iqtree.cpp:2026`
- `tree/phylokernelnonrev.cpp:656`

For most of these the underlying model is unlikely to be `ModelSet`, but it's worth a quick audit ‚Äî especially the simulator (`alisimulator.cpp:660, 691`), which would now produce simulations using pattern-0 frequencies if a user runs `--alisim` with a PMSF model.

**Suggested mitigation (one-line):** either change the default arg to `int mixture = -1` so the no-arg call still gets the averaged behaviour, or audit the 20 call sites and add explicit `-1` where averaged is intended. The former is safer and consistent with the comment in the new function: `// default: return the +F freqs across all patterns`.

## CI

5 of 7 platforms passing. Failing:
- Mac OS x86-64
- Linux x86-64 (22, latest, true)

The failures are not platforms I'd expect to be sensitive to this change (Mac/Linux x86_64 should behave the same as ARM64, which passes). Possibilities:
- Test fixture comparing logged pinv value: the bugfix legitimately changes the optimal pinv from 0.0 to something positive for any PMSF+I test case, which is **exactly the intent**. If a regression test was pinning the buggy value, the test needs updating, not the code.
- Spurious infrastructure failure.

Please pull the CI logs and confirm whether the diff in any failing test is just `pinv: 0 ‚Üí 0.something` and a corresponding logL improvement ‚Äî if so, that's evidence the fix works and the baseline needs refreshing.

## Style / minor

- `int state_freq[nstates]` at `phylotreesse.cpp:566` is a VLA (Clang/GCC extension). Same as the old code ‚Äî no regression. If you eventually switch to `std::vector<double>` to silence the VLA warning, do it project-wide; don't single this out.
- `ASSERT(astate <= 14)` for DNA ambiguity (line 597) and `ASSERT(astate <= 2)` for protein (line 605) are good defensive guards that match the IUPAC encoding.

## Verdict

**Approve-with-comments.** The fix is correct, well-scoped, and addresses the reported bug at its root. The only thing I'd ask before merging:
1. Either change the default for `ModelSet::getStateFrequency` from `mixture = 0` to `mixture = -1`, **or** confirm by audit that none of the 20 no-arg callers above behave badly on a ModelSet input (especially the AliSim path).
2. Investigate the 2 CI failures ‚Äî most likely a pinned-value test that needs updating because the fix legitimately produces a different (correct) pinv.

bqminh (approved):

StefanFlaumberg:
@bqminh,
Here Claude found the right problem (the `mixture` parameter being defaulted to 0 instead of -1), but proposed a wrong solution. The one-line fix in the signature of the `ModelSet` function actually leads to even more confusion: since it is a virtual function, every its call from the base class (`ModelMarkov` or `ModelSubst`) will use the default value from the signature of the base function, i.e. `mixture = 0`. So despite your fix, in most cases 0 will still be used -- and this would come as a surprise for anyone trying to fix any related bug.
I knew about the problem, but since it requires fixing all such signatures at once and since the function is never really used in a way that would lead to a bug, I opted not to touch it. Unfortunately, you added the proposed fix before I could even notice the review (in the future, please give me some time to reply first).
Now I'm going to propose a comprehensive fix in a separate PR.

Best regards,
Stefan

PR #170, opened 2026-05-26
Update libomp.a built with older targets (10 and 11 rather than 26)

trongnhanuit:
Update libomp.a built with older targets (10 for Intel chip and 11 for Arm chip, rather than 26) to avoid compilation warnings


bqminh (approved):
compile well now for mac

PR #173, opened 2026-06-09
Fix two bugs associated with the -mem option

StefanFlaumberg:
This pull request introduces the following changes:
- Fix bug 1 from iqtree/iqtree2#431 by setting `max_lh_slots` independently of `central_partial_lh` to cover the cases where `central_partial_lh` is copied as in `ModelMixture::optimizeWithEM()`, which was causing the bug.
- Fix bug 2 therefrom by capping `min_lh_slots` with `max_lh_slots`, as the former number can be overestimated for small trees with less than 5 leaves.
- Also, in `PhyloTree::getMemoryRequired()`, correct the memory estimates for UFBoot and site-specific models, make type conversions more explicit, and add more comments.

### DIFF
--- a/tree/phylotree.cpp
+++ b/tree/phylotree.cpp
@@ -903,7 +903,6 @@ size_t PhyloTree::getBufferPartialLhSize() {
 }
 
 void PhyloTree::initializeAllPartialLh() {
-    int index, indexlh;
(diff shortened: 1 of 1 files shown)


bqminh (approved):
good

PR #174, opened 2026-06-15
Mixlen refactoring: Make GHOST compatible with PMSF

StefanFlaumberg:
This pull request implements the GHOST rate heterogeneity model for the PMSF model, a long-awaited feature requested several times before:  https://github.com/iqtree/iqtree2/pull/391#issuecomment-2601064828 and iqtree/iqtree2#459.

**Implementation details:**
- Fix a latent bug in the `PhyloTreeMixlen::computeFuncDerv()` function by switching the condition for using the base class-version function from `if (cur_mixture < 0)` to `if (initializing_mixlen)` and by asserting that `cur_mixture` is set. The rationale is that the base class-version function should be called only when the heterotachy model is not used yet, i.e. when `initializing_mixlen` is true. Otherwise, it would, in turn, call `PhyloTree::computeLikelihoodDerv(Generic)SIMD()` that would run in the mixlen mode, thus returning partial derivatives instead of the single ordinary derivative. Also, the unused code duplicating the one from `phylokernelnew.h` is now removed from this function.
- In the `PhyloTree::computeLikelihoodDerv(Generic)SIMD()` function, in the `if (isMixlen())` branch, duplicate the `if (SITE_MODEL)` block of the `else` branch with the only difference that the length is considered category-specific here. Make the two code branches more similar.
- Change `->length` to `->getLength(c)` throughout the `phylokernelnew.h` code, most importantly in the `if (SITE_MODEL)` blocks, to ensure that all computations are compatible with the mixlen mode.
- For now, the PMSF model can be run only under the BFGS optimization of heterotachous branch lengths. Users running the default command with the EM algorithm will get the annoying error message advising switching to the BFGS algorithm. This inconvenience is intended to make the users informed that they have to run the optimization algorithm different from the one used in the original GHOST model paper. However, in another upcoming pull request, I am going to make the EM optimization work as well, so the restriction should be lifted soon.

**A simple test to check the new likelihood calculations:**
Run cmd1: `iqtree3 --seed 123 -nt 1 -s aln.faa -m LG+F+H2 -optlen BFGS --keep-ident -pre test_LG_F_H`
and cmd2: `iqtree3 --seed 123 -nt 1 -s aln.faa -m LG+H2 -fs empty.sitefreq -optlen BFGS --keep-ident -pre test_LG_SSF_H`,
where `empty.sitefreq` is literally an empty text file:
```
$ wc empty.sitefreq
0 0 0 empty.sitefreq
```
This two runs should have exactly the same likelihood, tree, and parameter estimates, since `-fs empty.sitefreq` activates a `+SSF` model that uses the `+F` state probability vector for each alignment pattern, which simulates the simpler `+F` model yet using site-specific computations.

### DIFF
--- a/tree/phylokernelnew.h
+++ b/tree/phylokernelnew.h
@@ -5,16 +5,11 @@
  *  Created on: Sept 23, 2016
  *      Author: minh
  */
-
-
 #if !defined(PHYLOKERNELNEW_H_) || !defined(PHYLOKERNELNEW_STATE_H_)
-
-#define JAMES_VERSION 1
-
 #ifdef KERNEL_FIX_STATES
-#   define PHYLOKERNELNEW_STATE_H_
+#define PHYLOKERNELNEW_STATE_H_
 #else
-#   define PHYLOKERNELNEW_H_
+#define PHYLOKERNELNEW_H_
 #endif
 
 #include "phylotree.h"
@@ -27,9 +22,11 @@
 
 /*******************************************************
  *
- * Helper function for vectors and matrix multiplication
+ * Helper functions for vector and matrix multiplication
  *
  ******************************************************/
+
+#define JAMES_VERSION 1
 
 /**
     sum of elments of a vector:
@@ -883,14 +880,14 @@ inline void scaleLikelihood(VectorClass &lh_max, double *invar, double *dad_part
     }
 }*/
 
+#undef JAMES_VERSION
 
 /*******************************************************
  *
- * Helper function to pre-compute traversal information
+ * Helper functions to pre-compute traversal information
  * and buffer to transition matrix
  *
  ******************************************************/
-
 
 #ifdef KERNEL_FIX_STATES
 template<class VectorClass, const int nstates>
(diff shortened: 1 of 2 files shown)


StefanFlaumberg:
@bqminh,
I was hoping it'd be included in the v3.1.3 release...

bqminh:
I'm reviewing this now ... it's touching the kernel, so need be to thorough

bqminh (approved):
looks great. thanks. Suggestion: for future pull requests, better separate into smaller PRs, instead of combining many into one.

bqminh:
It's now released in v3.1.3

StefanFlaumberg:
> It's now released in v3.1.3

@bqminh, thank you!

StefanFlaumberg:
Hi @bqminh,

The v3.1.3 precompiled binaries exactly reproduce the behaviour of the v3.1.2 binaries when run with the PMSF+GHOST model: print out the wrong initial likelihood and immediately crash (thanks to @kr3118 for pointing this out to me).
I've tested only the iqtree-3.1.3-Linux-intel package, but others may have this problem as well. On the other hand, the binary compiled from the v3.1.3 source code runs fine. I guess you must somehow have added wrong binaries for the release.

Could you recompile and update the binaries if it is possible? Thanks!

Best,
Stefan

PR #175, opened 2026-06-17
Fix EM mixture-frequency optimization

thomaskf:
Two fixes in `ModelMixture::optimizeWithEM` (used by `-mfopt -optalg_qmix EM`). The issues were reported by Ryo.

1. _EM made the result worse instead of better_ (the final log-likelihood was lower than the start). The  E-step used old partial-likelihood values that were not updated for the current model, so the per-class weights were wrong. Fix: call `clearAllPartialLH()` before each `computePatternLhCat` that builds the weights, so the values are recomputed. 

2. _Crash when `-mem` is used_.  Fix: copy `max_lh_slots` to the inner tree.

- Tested on LG+C10+G4 (20 taxa, 5000 sites): EM now improves correctly, runs with `-mem 4G` and `-mem 1G` without crashing, and  the non-`-mem` and BFGS runs are unchanged.

### DIFF
--- a/model/modelmixture.cpp
+++ b/model/modelmixture.cpp
@@ -4177,6 +4177,7 @@ double ModelMixture::optimizeWithEM(double gradient_epsilon) {
     // EM algorithm loop described in Wang, Li, Susko, and Roger (2008)
     for (step = 0; step < optimize_steps; step++) {
         // first compute _pattern_lh_cat
+        phylo_tree->clearAllPartialLH();
         score = phylo_tree->computePatternLhCat(WSL_MIXTURE);
 
         if (score < prev_score + gradient_epsilon)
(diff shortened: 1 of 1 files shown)


StefanFlaumberg:
Hi @thomaskf,

The bug causing the `-mem` crash has already been fixed by yet not merged #173 in a more safe and general way: the `max_lh_slots` member, if zero, is always computed by a `initializeAllPartialLh()` call. In `ModelMixture::optimizeWithEM()`, this call happens at the end of the across-submodel for-loop.
I think having `max_lh_slots` always checked and set automatically is a better solution than copying it manually because if a similar problem occurs in some similar place in the code where a temporary copy of the tree is used, it will be already solved.

Best regards,
Stefan

thomaskf:
Thanks @StefanFlaumberg 
Since the `-mem` issue is already addressed in your PR, there is no need to fix it here as well. I have therefore removed that fix, leaving only the fix for the EM optimisation issue.

bqminh (approved):
all good

PR #182, opened 2026-07-07
Mixlen refactoring: Remove dependency on ModelFactoryMixlen and cppoptlib, add support for the -ft option

StefanFlaumberg:
This pull request simplifies the GHOST rate model implementation and allows using mixture models with GHOST rates for PMSF profile estimation on a guide tree. Specifically, it introduces the following changes:
- Merge the `ModelFactoryMixlen` class into the `ModelFactory` class. The former class was just adding unnecessary complexity, its functionality is now seamlessly included into the `ModelFactory` class.
- Remove dependency of the `PhyloTreeMixlen` class on the `cppoptlib` library and simplify initialization of `PhyloTreeMixlen` objects. The library has never been compiled and used, its version has long been deprecated, and if needed someday (very unlikely), it can be reimplemented without inheriting from `PhyloTreeMixlen`. The library required the number of heterotachy categories being set by the `PhyloTreeMixlen()` constructor (and used the ill-defined `-mixlen` option for this), this largely contradicted the general logic that implies that the number of categories should be passed within the model string (as `-m "...+Hn"`).
Now the logic is respected: the `PhyloTreeMixlen()` constructor takes only the alignment object as an argument, the number of heterotachy categories is set exclusively by the `PhyloTreeMixlen::setModelFactory()` function as given in the model string, and if a tree is read from a file with not enough or with extra sets of branch lengths, it is corrected by the `PhyloTreeMixlen::initializeMixBranches()` function as before.
- With the new `PhyloTreeMixlen` class initialization rules (see above), `mixlen` is always 1 for a newly constructed object and is always higher than 1 after setting a `RateHeterotachy` site rate with `setModelFactory()`, it can never be 0. Also ASSERTs are added to ensure that a model factory provides a `RateHeterotachy` site rate (or a nullptr) and that a `RateHeterotachy` site rate uses a `PhyloTreeMixlen` tree.
- Fix `PhyloTree::getCurMixture()` to return -1 instead of 0. The old output likely didn't affect anything for now, but clearly was a latent bug.
- Fix and simplify switching to `PhyloTreeMixlen` in `AliSimulator::initializeIQTreeFromTreeFile()`. The fix comes from declaring `IQTree *&current_tree` as a reference to a pointer, not just a pointer in the supertree case -- this way the subtrees are reset, not just deleted in the supertree vector.
- Use the `IQTree` or `PhyloTreeMixlen` object instead of the `PhyloTree` object for the guide tree in the `computeSiteFrequencyModel()` function of `phyloanalysis.cpp` to extend the `-ft` option to the mixture models that use heterotachy rates. This makes the fix for the iqtree/iqtree2#459 issue complete.
Note `params.min_branch_length` being temporarily set in the function: this is needed in the general case, but the need is especially pronounced for the heterotachy rate model, which can be using very short branch lengths (without the minimal length, this results in numerical underflows).
- Minor: in the `runStandardBootstrap()` function of `phyloanalysis.cpp`, change the indices of the `.bootlh` file lines from always zero to using `sample`. This is the output of the `-wtl` option; it looks pretty useless, for the likelihoods are not phylogenetic, but polynomial, but let it at least have the correct indexing now.

These changes are intended not to affect tree and model optimization. The optimized likelihoods should be the same as in the upstream version wherever they can be compared.


StefanFlaumberg:
**The remaining problems:**
There are many. The `PhyloTreeMixlen::retoreCheckpoint()` function doesn't work at the beginning of the analysis, when mixlen = 1, due to checkpoint key mismatch. The IQTree -> PhyloTreeMixlen switching block, now placed in `startTreeReconstruction()`, is enough for ModelFinder, but not for MixtureFinder, which runs on tree copies, so the heterotachy rate model remains incompatible with MixtureFinder. The `treeLengths()` function accumulates values across successive calls, but I haven't touched this, as the fix affects analysis results. The logic behind the `readTreeString()` function setting `relative_treelen` while `relative_treelen` can also be initialized from a temp free rate model is quite complicated and a bit circular.
I opted not to fix these problems here to preserve the original algorithm and likelihoods. I will try to address them in the following PRs.

StefanFlaumberg:
**On the new changes:**
- Rename `getMixlen()` to `getNMixlen()`.
- Merge the `save/restoreBranchLengths()` functions of `PhyloSuperTreeUnlinked` into the respective base functions of `PhyloSuperTree`. For both `PhyloTree` and `PhyloSuperTree` classes, make `lenvec` checking/resizing more robust in these functions (too small vectors now get properly resized).
- Fix `PhyloTree::getNBranchParameters()` to properly account for mixlen and rooted trees. Previously, the return of the function was augmented in `ModelFactory::getNParameters()`, but this gave wrong results for rooted trees and branch scale optimization.

StefanFlaumberg:
@bqminh,
The conflict arises due to one newline having been modified in the last PR. If you are not going to propose any changes here, please try to resolve the conflict right within the merge commit if possible (I'd rather not add a dummy newline-removing commit).

bqminh:
This PR has some conflicts with #193. Since 193 changed the code much less, I will merge #193 first, and then this one.

StefanFlaumberg:
@bqminh,
Very strange, actually, as I can find no conflicts at all... There is the conflict in `alisimulator.cpp` with the `master` branch (but not the `mutsel2` branch!), you can fix it manually within the merging commit by simply removing that extra blank line, just as I wrote above. Then `mutsel2` can be merged without any conflicts. I've just tried it myself. The changes seem to be completely orthogonal in these two PRs.

But otherwise I don't mind.

bqminh:
There is now some conflict... (after I merged mutsel and other PRs). Can you have a look? I'll prioritise to merge this PR next

StefanFlaumberg:
@bqminh,
The merge conflicts are resolved now.

bqminh:
will do. We have to release 3.1.4 now for a publication

StefanFlaumberg:
> We have to release 3.1.4 now for a publication

@bqminh, I understand and see the hurry...
I saw that you've only pre-released v3.1.4 for now. If you consider this still possible somehow, it'd be great if the PR could be included in the final v3.1.4 release, since it implements the GHOST + `-ft` option combination, which has been requested and awaited by other users since quite a while ago

simulator/alisimulator.cpp
```diff
@@ -151,29 +154,8 @@ void AliSimulator::initializeIQTreeFromTreeFile()
...
-                IQTree* new_tree = new PhyloTreeMixlen(current_tree->aln, current_tree->getRate()->getNRate());
-                
-                // delete the old tree
-                delete current_tree;
-                
-                // set the new PhyloTreeMixlen to the new tree
-                current_tree = new_tree;
-                
-                // re-load the tree/branch-lengths from the file
-                current_tree->IQTree::readTree(params->user_file, is_rooted, tree_line_index);
```
bqminh, line 171: why is the line current_tree->IQTree::readTree(params->user_file, is_rooted, tree_line_index); removed?

bqminh: wait, it is in line 141. all good

tree/phylotreemixlen.cpp
```diff
@@ -5,121 +5,70 @@
...
         startCheckpoint();
-        if (this->relative_treelen.size() > 0) {
-            ASSERT(mixlen == this->relative_treelen.size());
-            double relative_treelen[mixlen];
-            for (int i = 0; i < mixlen; i++)
-                relative_treelen[i] = this->relative_treelen[i];
-            CKP_ARRAY_SAVE(mixlen, relative_treelen);
+        if (relative_treelen.size() > 0) {
+            ASSERT(mixlen == relative_treelen.size());
+            CKP_ARRAY_SAVE(mixlen, &relative_treelen[0]);
```
bqminh, line 37: Here you actually changed the checkpoint file! the key is very different now. Have a look at the macro CKP_ARRAY_SAVE. If you view the ckp.gz file, the line looks like:

 &relative_treelen[0]: 527.8043991, 568.0718784

whereas normally it looks like:

 relative_treelen: 527.8043991, 568.0718784

bqminh, line 37: Here you actually changed the checkpoint file! the key is very different now. Have a look at the macro CKP_ARRAY_SAVE. If you view the ckp.gz file, the line looks like:

 &relative_treelen[0]: 527.8043991, 568.0718784

whereas normally it looks like:

 relative_treelen: 527.8043991, 568.0718784

bqminh, line 37: Here you actually changed the checkpoint file! the key is very different now. Have a look at the macro CKP_ARRAY_SAVE. If you view the ckp.gz file, the line looks like:

 &relative_treelen[0]: 527.8043991, 568.0718784

whereas normally it looks like:

 relative_treelen: 527.8043991, 568.0718784

bqminh: i've reverted these changes locally and push later. Also commented about that

tree/phylotreemixlen.cpp
```diff
@@ -5,121 +5,70 @@
...
+    if (mixlen > 1) {
         startCheckpoint();
-        double relative_treelen[mixlen];
-        if (CKP_ARRAY_RESTORE(mixlen, relative_treelen)) {
-            this->relative_treelen.resize(mixlen);
-            for (int i = 0; i < mixlen; i++)
-                this->relative_treelen[i] = relative_treelen[i];
+        DoubleVector restored_relative_treelen(mixlen, 0.0);
+        if (CKP_ARRAY_RESTORE(mixlen, &restored_relative_treelen[0])) {
+            relative_treelen = restored_relative_treelen;
```
bqminh, line 49: Same problem here

bqminh (dismissed):
All good, except for one comment, which I will fix. Thanks for the contributions!
May I please suggest that: for future release can you please make several smaller PR? It's just easier to review, and make it less error prone.

bqminh (dismissed):

StefanFlaumberg:
@bqminh,
I see, that was quite a dumb mistake of mine. Thank you for the review and for fixing it!

bqminh:
No worries, actually i am impressed with your understanding of the large IQ-TREE code base (I think we have >200,000 lines of code). My job is to review carefully...

PR #191, opened 2026-07-23
Fix AliSim crashing when using GHOST with a partitioned/mixed-datatype models

trongnhanuit:
This pull request fixes a crash when AliSim simulates alignments using GHOST with partitioned/mixed-datatype models because of a deleted tree pointer. This bug was reported in [the iqtree2 repo](https://github.com/iqtree/iqtree2/issues/537).
Solution: set the pointer to the correct (newly-created) PhyloTreeMixlen tree.

### DIFF
--- a/simulator/alisimulator.cpp
+++ b/simulator/alisimulator.cpp
@@ -166,10 +166,15 @@ void AliSimulator::initializeIQTreeFromTreeFile()
                 
                 // set the new PhyloTreeMixlen to the new tree
                 current_tree = new_tree;
-                
+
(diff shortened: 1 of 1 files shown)


StefanFlaumberg:
Hi @trongnhanuit,

This issue has already been solved in #182 in a much more general way. No need for a duplicate solution.

Best,
Stefan

bqminh (approved):
I understand that this issue is also solved in another PR #182 , but since this one is much easier tor review, I'll merge it first.

PR #197, opened 2026-08-04
Bug fix: uniformly select indel positions across only non-gap sites

trongnhanuit:
**Issue:** When an indel event occurs, AliSim picks a random site across the entire sequence, including both gapped and non-gapped sites. If it lands on a gapped site, AliSim shifts forward to the nearest non-gapped site and uses that instead. This makes site selection non-uniform - sites right after existing gaps are more likely to be picked than others. 

**Solution:** AliSim now selects uniformly among only the non-gapped sites.


bqminh (approved):
all good

PR #202, opened 2026-08-11
PartitionFinder-mAIC improvement

HuaiyanRen:
speed up PartitionFinder-mAIC by:
- remove copyTree()
- change site-lh to pattern-lh
- cache the lh vector of partition A and model B in the partitioning scheme and reuse (the most important for runtime)
- flat the two-layer loop (j for partitions, k for models) to one loop (for (j,k) pair)

move the final mAIC computation to before final time stamp.

set mAIC as default option for PartitionFinder.


bqminh (approved):
all good

PR #209, opened 2026-09-08
Use different memory thresholds for different platforms in ci workflow

thomaskf:
- Use different memory thresholds for different platforms in ci workflow
- Clean up incomplete message for ModelFinder


bqminh (approved):
all good

PR #212, opened 2026-09-09
use different thresholds for different platforms in time testing

thomaskf:
Use different thresholds for different platforms in running time testing for ci workflow


bqminh (approved):
all look good thanks

PR #213, opened 2026-09-09
Fix the CI workflow

thomaskf:
- Moved the two new commands to the end of the list, so the other tests line up with their thresholds again.
- The workflow now skips any test that has no threshold.
- If a time or memory test fails, that single test is run again. It is reported as a failure only if it fails the second time as well. This avoids failures caused by a slow or busy CI machine.

### DIFF
--- a/test_scripts/remeasure.ps1
+++ b/test_scripts/remeasure.ps1
@@ -1,0 +1,54 @@
+# Re-run a single benchmark command and return @{ Time = <seconds>; Mem = <MB> }.
+#
+# Used by verify_memory.ps1 / verify_runtime.ps1 to retry a command that breached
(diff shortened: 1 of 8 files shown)


bqminh (approved):
Looks great, thanks for the CI fix. Moreover, I'll remove all these new tests, because:

*  The CI is not meant to be an extensive test suite. For extensive testing, developers need to do it internally, and show the results via other means. Some test files may be uploaded to the PR to help reproduce the intended behaviour.
* Instead, the CI checks a few most common use cases, making sure the behaviour is the same between IQTREE 2 and 3. So of course when there is new option/feature, this is not the place to test.
* The CI auto builds the binaries, so that some users may early beta-test the versions before official release.

Issue #196, opened 2026-08-01
Jackknife proportion is inverted


digitalwright:
Please correct me if I am wrong, but I believe the jack-prop argument is inverted.

It is described as:
`--jack-prop NUM      Subsampling proportion for jackknife (default: 0.5)`

In the tools.cpp, the code uses it as follows:
```
int total = floor((1.0 - Params::getInstance().jackknife_prop)*n);
if (total <= 0)
    outError("Jackknife sample size is zero");
```

If `jack-prop` is 1 (defined as subsampling all the data [not subsampling]), then total would be zero and no sites would be selected (throwing an error). If `jack-prop` is 0 (defined as sampling no data), then total would be n and all sites would be selected.

I might be thinking about this backwards, but I believe jack-prop should be defined as:
`--jack-prop NUM      Deletion proportion for jackknife (default: 0.5)`

Of course, this has no effect at the default value since 1 - 0.5 = 0.5.

Most importantly, thank you for maintaining the program and taking a look at this report.

bqminh:
You are right, it's the fraction of data to remove with this option. I.e., 0.1 means to remove 10% of the sites and keep 90%. The wording of the help usage is not correct. We'll change that. Thanks.

PR #214, opened 2026-09-09
SH-aLRT: draw RELL replicates as bootstrap samples whatever -j/-J selected (#198)

### LINKED ISSUE #198: SH-aLRT changes when using jackknife
I noticed an unexpected discrepancy in SH-aLRT values when using jackknife versus bootstrap.

Attached is an example input file and output trees using commands:

```
iqtree3  -s M100_DNA_1.fas.gz --ufjack 1000 -alrt 1000 -T AUTO -ntmax 16 --seed 123 --seqtype DNA -m TEST
iqtree3  -s M100_DNA_1.fas.gz -bb 1000 -alrt 1000 -T AUTO -ntmax 16 --seed 123 --seqtype DNA -m TEST
```

Note the only difference between the two commands is the use of UFBoot or UFJack. Yet, the SH-aLRT values in the output trees are very different, and I do not know why that would be the case.

Further tests suggest SH-aLRT is providing lower resolution values when using any variant of jackknife (-J/-j) relative to any variant of bootstrap (-b/-bb). This occurs across many different input sequence sets.

I wanted to bring this to your attention in case it is unexpected behavior. Thank you for maintaining the code and reviewing this report.

[M100_DNA_1.fas.gz](https://github.com/user-attachments/files/30697778/M100_DNA_1.fas.gz)
[UFBoot.txt](https://github.com/user-attachments/files/30697797/UFBoot.txt)
[UFJack.txt](https://github.com/user-attachments/files/30697796/UFJack.txt)

cindykrafft:
Fixes #198.

`PhyloTree::resampleLh()` builds the RELL replicates of the SH-aLRT test (and of `-lbp`) with
`Alignment::createBootstrapAlignment()`, which follows the global `--jack-prop` setting. With `-j`
or `-J` every replicate was therefore a delete-half jackknife sample: a sum over half the sites,
whose expectation is half the full-alignment log-likelihood. The statistics `lh_new - lh` in
`testOneBranch()` were then no longer centred, and the support of every branch moved. On
`test_scripts/test_data/turtle.fa` with the test seed:

| run | SH-aLRT values on the same ML tree |
|---|---|
| `-B 1000 -alrt 1000` | 98.4 89.8 98.2 61.7 47.6 47.9 98 100 81.1 |
| `-J 1000 -alrt 1000` (before) | 100 100 100 91.7 86 81.5 100 100 99.7 |
| `-J 1000 -alrt 1000` (after) | 98.4 89.8 98.2 61.7 47.6 47.9 98 100 81.1 |

An independent numpy implementation of the `testOneBranch()` rule on IQ-TREE's own site
log-likelihoods gives the same picture on a simulated 24-taxon alignment: bootstrap RELL 60.3 /
54.6 / 74.6 / 85.8 on four branches, delete-half jackknife RELL 74.0 / 82.9 / 97.6 / 100.0, and
the jackknife pseudo-value (weights × 2, expectation restored) 57.5 / 53.8 / 74.0 / 85.1 — the
inflation is the missing centring, not sampling noise
(https://github.com/cindykrafft/mytochondria/tree/main/audits/iqtree/issue-fixes/198-sh-alrt-jackknife).

Change: `random_resampling()` and `createBootstrapAlignment(int*, ...)` (Alignment and
SuperAlignment) take an optional jackknife proportion, negative by default meaning "the global
setting", and `resampleLh()` passes 0 so the SH-aLRT replicates are bootstrap samples as in
Guindon et al. (2010) and Adachi & Hasegawa (1996). The tree-search resampling of `-j`/`-J` is
untouched (UFJack and UFBoot values are bit-identical before and after), as are the
`GENE`/`GENESITE`/`SCALE=` paths. If you would rather keep jackknife replicates and rescale them
(`lh_new / (1 - jack_prop)`), that is a two-line variant of the same change; I went with the
definition of the test.

Test: `test_iqtree.sh`/`.ps1` run `turtle.fa` with `-B 1000 -alrt 1000` and with `-J 1000 -alrt
1000` and fail if the SH-aLRT labels differ (they do on `master`); two `expect_ans.txt` rows for
the log-likelihoods. The full `test_iqtree.sh` + `verify_results.sh` pass locally with the change.

Not changed here, same pattern: the RELL replicates of the tree topology tests (`--test`, KH/SH
and the weighted variants in `treetesting.cpp`) also follow `--jack-prop`, and the AU test stops
with "Unsupported jackknife with SCALE=" when `-j`/`-J` is given. Happy to send that as a
follow-up if you agree it should behave the same way. #196 (the `--jack-prop` wording) already
has its fix on `master` (b8ef4238) and could be closed.

---
_Generated by [Claude Code](https://claude.ai/code)_


bqminh (changes requested):
Can you remove the CI stuffs? This is not really the intension of the CI.

The CI is not really meant to be extensive test suite. It's only to check a few most popular use cases by users, that should mimic functionality of IQ-TREE v2. We also want to build binaries, so that some users can download and early beta test it, before releasing. The CI has to run quick enough, right now keeping ~20 min. We don't want to make it run longer.

If you want more testing, then this needs to be done with other means. You can post some input files and commands to reproduce it with the PR, if you want.

cindykrafft:
Understood! I have removed the `test_scripts` changes
(`test_iqtree.sh`, `test_iqtree.ps1`, `expect_ans.txt`) and force-pushed; the branch is now the
six-file code change only, nothing that lengthens the CI run.

Here is the reproduction instead. It needs no extra input files: it uses `turtle.fa`, which is
already in the repository, and two runs of about a minute each.

```
iqtree3 -s test_scripts/test_data/turtle.fa -B 1000 -alrt 1000 --prefix turtle.alrt.boot -T 1 -seed 73073
iqtree3 -s test_scripts/test_data/turtle.fa -J 1000 -alrt 1000 --prefix turtle.alrt.jack -T 1 -seed 73073

grep -o ')[0-9.]*/' turtle.alrt.boot.treefile   # SH-aLRT values, UFBoot run
grep -o ')[0-9.]*/' turtle.alrt.jack.treefile   # SH-aLRT values, UFJack run
```

Both runs find the same ML tree with the same log-likelihood (-5370.3730), so the SH-aLRT values
should be the same up to the replicate noise of the test. On master they are not:

| run | SH-aLRT on the nine internal branches |
|---|---|
| `-B 1000 -alrt 1000` | 98.4 89.8 98.2 61.7 47.6 47.9 98 100 81.1 |
| `-J 1000 -alrt 1000`, master | 100 100 100 91.7 86 81.5 100 100 99.7 |
| `-J 1000 -alrt 1000`, this branch | 98.4 89.8 98.2 61.7 47.6 47.9 98 100 81.1 |

The UFBoot and UFJack support values themselves are unchanged by the patch, on this data
bit-identical before and after; only the SH-aLRT column moves.

A second check on simulated data (AliSim, 24 taxa, JC, 800 sites, `-alrt 1000` alone / with
`-bb 1000` / with `--ufjack 1000`) and an independent numpy re-implementation of the
`testOneBranch()` rule, which separates the missing centring from Monte-Carlo noise, are in the
kit if they are useful:
https://github.com/cindykrafft/mytochondria/tree/main/audits/iqtree/issue-fixes/198-sh-alrt-jackknife

Happy to run anything else you would like to see before merging.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #215, opened 2026-09-10
Remove -ft AUTO and match benchmark thresholds by identifier

thomaskf:
## 1. Remove the process-spawning code and disable -ft AUTO

Removes utils/self_process.h and the spawn call, and disables -ft AUTO at argument parsing. The spawned process only received the alignment, a fixed LG+F+G4 model, the thread count and a prefix, so all other user parameters were ignored. Users should infer a guide tree themselves and pass it to -ft.

The removed header also required the C++17 <filesystem> library, which broke compilation on Windows with older compilers. That dependency is now gone.

## 2. Match benchmark thresholds by identifier, not by position

In the CI workflow we now use the name of the .iqtree file that each command produces as the identifier, the same form already used by expect_ans.txt. The threshold is matched by this identifier instead of by position, so a command added in the middle of the list is still verified against the correct threshold.

A command with no row is skipped with a message saying which file to edit, and a row with no command is reported as a warning. Neither fails the build.

The two files are renamed to expect_memory.txt and expect_runtime.txt. Their lines are now much shorter, because we keep only the short .iqtree identifier instead of the whole command, so the files are easier to maintain.


bqminh (approved):
looks good thank you!

bqminh:
@benjamin-lieser Thomas had to remove process spawn stuff due to the reason of incompatibility with old Windows. However as I pointed out earlier, if you wanna have this option back, have a look at places where we call runTreeReconstruction() function.

